PR #31, opened 2012-08-10
allow value-concatenating an object to a list (master)

### LINKED ISSUE #24: allow value-concatenating an object to each list element
0) this is a feature request

1) use case: 
- imaging you have network service with some 100 end points;
- each end point has some 30 parameters, mostly common to all points;
- may be  2 to 5 of the parameters are to be overridden by each point;

2) the way I understand current mechanisms of HOCON 
https://github.com/typesafehub/config/blob/master/HOCON.md
this would make this definition very verbose;

3) how about new/proposed `object$builder` syntax:

```
//
// the list definition below will be looking for this builder
//
netty_point_list$builder {
   id = "you-must-override"
   localAddress = "1.2.3.4:12345"
   remoteAddress = "you-must-override"
   packetTTL = 5
   pipeline = "default-pipeline"
   pipeline_param_0 = 123
   // 20 more parameters here; mostly come from include  or variable substitution
}

//
// netty_point_list will be looking for optional netty_point_list$builder
// to initialize default list items parameters before override
//
netty_point_list [
   { id = "service-1", remoteAddress = "1.1.1.1:12345", pipeline = "advanced" }
   { id = "service-2", remoteAddress = "1.1.1.2:12345", localAddress = "3.3.3.3:12345" }
   // 100 more service definitions with little variations on the builder-defined parameters
]
```

4) am I missing some HOCON feature or what? :-)

carrot-garden:
allow value-concatenating an object to a list by merging the object into each list element ( fixes #24)

### DIFF
--- a/HOCON.md
+++ b/HOCON.md
@@ -249,7 +249,7 @@ of at least three quotes ends the multi-line string, and any
 ### Value concatenation
 
 The value of an object field or array element may consist of
-multiple values which are combined. There are three kinds of value
+multiple values which are combined. There are 4 kinds of value
 concatenation:
 
  - if all the values are simple values (neither objects nor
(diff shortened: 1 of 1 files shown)


2m:
This has been open for a very long time and has had no activity lately. I am going to close this. Feel free to open another PR or continue the discussion in relevant tickets.

PR #169, opened 2014-06-10
Added support for YAML files.

derrickburns:
This pull request adds support for parsing YAML files using the SnakeYAML code. It needs to be tested.  I am not 100% sure that it works I do not completely understand the typesafe/config design. 

The reason for this modification is to make it easier to introduce HOCON into environments that use YAML.  With this change, a HOCON file can include a YAML file. At least that is the intent! :)

### DIFF
--- a/config/build.sbt
+++ b/config/build.sbt
@@ -18,6 +18,8 @@ libraryDependencies += "net.liftweb" %% "lift-json" % "2.5" % "test"
 
 libraryDependencies += "com.novocode" % "junit-interface" % "0.10-M4" % "test"
 
+libraryDependencies += "org.yaml" % "snakeyaml" % "1.12"
+
 externalResolvers += "Scala Tools Snapshots" at "http://scala-tools.org/repo-snapshots/"
 
 seq(findbugsSettings : _*)
--- a/config/src/main/java/com/typesafe/config/ConfigSyntax.java
+++ b/config/src/main/java/com/typesafe/config/ConfigSyntax.java
@@ -32,5 +32,10 @@ public enum ConfigSyntax {
      * >Java properties</a> format. Associated with the <code>.properties</code>
      * file extension and <code>text/x-java-properties</code> Content-Type.
      */
-    PROPERTIES;
+    PROPERTIES,
+    /**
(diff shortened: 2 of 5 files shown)


derrickburns:
The snakeyml jar is only 274KB. Is this size of this jar a big deal?

havocp:
it isn't about size it's about coupling - what if an app already uses YAML elsewhere, what if dependency jars change ABI, what if dependency jars themselves acquire dependencies, and how hard is it to maintain config itself if we start adding new formats to the core.
That sort of concern.

viktorklang:
IMHO features like these should be solved as additional/external
integrations.

-- 
Cheers,
√

derrickburns:
If shade the YAML jar, then there are no issues.

havocp:
https://github.com/typesafehub/config/pull/243 introduced a public API to mess with `ConfigOrigin`. The way I'd like to see this feature proceed is that we file issues for missing public API, until we have public API complete enough to allow adding more formats from another jar.

davidkarlsen:
I would love to see this included. I've been looking all over for a java lib with HOCON and yaml include support - and here it is. What stops this from being merged?

havocp:
See comments above, want it to be an optional plugin.

davidkarlsen:
OK - so concretely - if the YAMLParser.java in this PR was a separate drop-in jar you would be ok with it? How about using a SPI, and discovering implementations via https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html ?

havocp:
The sequencing I'd suggest is to first do PRs for any missing public APIs needed to write a function taking a YAML character stream and returning a Config instance, so that functions like `Config parseYamlFile(File f)` or `Config parseYamlString(String s)` etc. can be implemented in a separate library. I'm not sure how many PRs there are here, maybe it's only one or two things, essentially replacing anything in YAMLParser.java that needs stuff from the `impl` package.

To prove this is solved, might want to actually implement said separate library (maintained in another repo, I think, it's logically its own thing).

Then in a separate PR, perhaps add the SPI/ServiceLoader feature, which would allow the yaml jar to hook in to ConfigFactory.load etc.

That seems fine to me anyway...

Issue #228, opened 2015-01-05
rendered keys of an object should be sorted alphanumerically


derrickburns:
When arrays are turned into JSON, they are represented as objects where the keys are numbers, e.g.

x {
  foo = [ "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k" ]
}

is equivalent to

x { foo {
  "0": "a"
  "1": "b"
  "2": "c"
  "3": "d"
  "4": "e"
  "5": "f"
  "6": "g"
  "7": "h"
  "8": "i"
  "9": "j"
  "10": "k"
} }

Unfortunately, when rendered this will appear as:

x { foo {
  "0": "a"
  "1": "b"
  "10": "k"
  "2": "c"
  "3": "d"
  "4": "e"
  "5": "f"
  "6": "g"
  "7": "h"
  "8": "i"
  "9": "j"
} }

The keys are sorted using the default sort.  They should be sorted using an alphanumerical comparator.

havocp:
This shouldn't happen - arrays that are arrays should remain so. The conversion that does happen is that if you have an object with numbers for field names, you can call Config.getList and have it treated as an array.

derrickburns:
But it does happen if you use an external file reader that flattens a file (YAML actually) into a map and then converts that map into a config.

It also happens it one uses numerical values for keys in an object

Regardless, the fix is trivial. Simply provide a comparator to the sort method that does this right thing. See my pull request on this.

havocp:
Agreed, I'm not against trying to fix this, just checking that arrays aren't converted to objects (that would be a much more serious bug if they were, and the solution would be different from formatting objects more nicely).

derrickburns:
I added code that is called when parseMap is used. This code will detect when maps actually encode arrays and to representing them as SimpleConfigLists instead of SimpleConfigObjects.  This lessens the need to sort map keys alphanumerically, although this is still a nice feature.

PR #229, opened 2015-01-05
Sort keys alphanumerically.

derrickburns:
Addresses #228.


typesafehub-validator:
Hi @derrickburns,

typesafehub-validator:
Hi @derrickburns,

havocp:
Thanks!

We have some extra things mixed in here - I think this PR should be the first commit only.

Unfortunately that first commit has a problem that it pulls in LGPL code, which would LGPL-ify the entire codebase by association. The Scala/Akka/Play stack has a rule that it does not have any (L)GPL dependencies. So we need a different implementation of that sort algorithm (and it cannot be based on or derived from the LGPL one, it needs to be a fresh from scratch implementation).

I think it might be possible to cheat somehow. For example, what if we took all the numeric keys and prefixed them with `0` padding to make them all the same length, so we would be sorting `"01", "02", "03"` etc., and then post-sort we strip the 0's back off? Obviously we'd only want to do this if the keys were all numeric and didn't already start with 0. But it would allow us to use the regular default string sort and still get things in the right order perhaps.

derrickburns:
What about this implementation:

https://github.com/sawano/alphanumeric-comparator

derrickburns:
I replaced the Comparator implementation with an Apache 2.0 licensed one.

This pull request also addresses #230.

typesafehub-validator:
Hi @derrickburns,

derrickburns:
I added code that is called when parseMap is used. This code will detect when maps actually encode arrays and to representing them as SimpleConfigLists instead of SimpleConfigObjects.  This lessens the need to sort map keys alphanumerically, although this is still a nice feature.

PR #232, opened 2015-01-07
Added alphanumeric sorting.

derrickburns:
Added alphanumeric sorting. Addresses #228.

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/SimpleConfigObject.java
+++ b/config/src/main/java/com/typesafe/config/impl/SimpleConfigObject.java
@@ -364,11 +364,15 @@ final class SimpleConfigObject extends AbstractConfigObject implements Serializa
         });
     }
 
+
     @Override
     protected void render(StringBuilder sb, int indent, boolean atRoot, ConfigRenderOptions options) {
         if (isEmpty()) {
             sb.append("{}");
         } else {
+            String[] keys = keySet().toArray(new String[size()]);
+            Arrays.sort(keys, new com.typesafe.config.impl.AlphanumericComparator());
+
             boolean outerBraces = options.getJson() || !atRoot;
 
             int innerIndent;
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @derrickburns,

havocp:
Thinking about this more, why not use a simpler comparator that also would give us copyright assignment. We can make these simplifying assumptions I believe:

 * we don't care about sorting strings that only start with a digit in a special way, we only care about all-digits strings
 * we don't care about floating point or negative numbers or hex or other "weird" numbers containing nondigits
 * we don't care about locale-aware sort, String.compareTo is fine
 * we are happy to sort all all-digits strings to the front of the list, so all-digits strings can always be "less than" any other string, ensuring the transitive property holds

So in sketched pseudocode something like this I think is sufficient?

```
 int compare(String a, String b) {
    boolean aDigits = isAllDigits(a)
    boolean bDigits = isAllDigits(b)
    if (aDigits && bDigits) {
       return Integer.compare(Integer.parseInt(a), Integer.parseInt(b));
    } else if (aDigits) {
       return -1;
    } else if (bDigits) {
       return 1;
    } else {
       return a.compareTo(b);
    }
 }
```

That's a lot less code and we'd have copyright assignment. Since this is just for debugging rendering, no need to go overboard....

config/src/main/java/com/typesafe/config/impl/SimpleConfigObject.java
```diff
@@ -364,11 +364,15 @@ public AbstractConfigValue modifyChild(String key, AbstractConfigValue v) {
...
 
+
     @Override
     protected void render(StringBuilder sb, int indent, boolean atRoot, ConfigRenderOptions options) {
         if (isEmpty()) {
             sb.append("{}");
         } else {
+            String[] keys = keySet().toArray(new String[size()]);
+            Arrays.sort(keys, new com.typesafe.config.impl.AlphanumericComparator());
+
```
havocp: if there's no reason to move this block of code around in the file, why don't we keep it where it was so the diff more accurately shows the change that was made?

derrickburns:
That's fine. It was faster me to find other working code than to write a new comparator from scratch.

config/src/main/java/com/typesafe/config/impl/SimpleConfigObject.java
derrickburns: Ok

typesafehub-validator:
Hi @derrickburns,

havocp:
See #266 for current PR

Issue #230, opened 2015-01-06
maps converted to Configs should detect encodings of lists


derrickburns:
When a Json list is encoded as a Map<String,String>, it is ambiguous whether the encoding represents a Json list of a Json map: if the keys are consecutive natural numbers starting from 0, then the encoding can be a map or a list.  However, it is unnatural to use such keys as keys to a map. Therefore, when decoding a Map<String,String> into a Config using parseMap, we should interpret such maps are lists.

PR #234, opened 2015-01-07
Represent as SimpleConfigLists maps that can be encoded as arrays. 

derrickburns:
Addresses #230.

### DIFF
--- a/config/src/main/java/com/typesafe/config/ConfigFactory.java
+++ b/config/src/main/java/com/typesafe/config/ConfigFactory.java
@@ -885,7 +885,7 @@ public final class ConfigFactory {
      */
     public static Config parseMap(Map<String, ? extends Object> values,
             String originDescription) {
-        return ConfigImpl.fromPathMap(values, originDescription).toConfig();
+        return ((ConfigObject) ConfigImpl.fromPathMap(values, originDescription)).toConfig();
     }
 
     /**
--- a/config/src/main/java/com/typesafe/config/ConfigFactory.java
+++ b/config/src/main/java/com/typesafe/config/ConfigFactory.java
@@ -885,7 +885,12 @@ public final class ConfigFactory {
      */
     public static Config parseMap(Map<String, ? extends Object> values,
             String originDescription) {
-        return ((ConfigObject) ConfigImpl.fromPathMap(values, originDescription)).toConfig();
+        return ((ConfigObject) ConfigImpl.fromPathMap(values, originDescription, false)).toConfig();
+    }
+
+    public static Config parseMapWithLists(Map<String, ? extends Object> values,
+                                  String originDescription) {
+        return ((ConfigObject) ConfigImpl.fromPathMap(values, originDescription, true)).toConfig();
     }
 
     /**
--- a/config/src/main/java/com/typesafe/config/ConfigParseOptions.java
+++ b/config/src/main/java/com/typesafe/config/ConfigParseOptions.java
@@ -26,18 +26,20 @@ public final class ConfigParseOptions {
     final boolean allowMissing;
     final ConfigIncluder includer;
     final ClassLoader classLoader;
+    final boolean inferLists;
 
     private ConfigParseOptions(ConfigSyntax syntax, String originDescription, boolean allowMissing,
-            ConfigIncluder includer, ClassLoader classLoader) {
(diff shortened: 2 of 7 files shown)


typesafehub-validator:
Hi @derrickburns,

havocp:
This breaks the current semantics and will break people / change ABI, I think. Right now the "parse a numeric-keyed object as a list" feature happens only at the time that you say you want a list, by calling getList(). It doesn't change the Config itself, only "casts" the value on demand. That way if you really want an object with numeric keys, you can still have one. The trouble with a heuristic like this done eagerly/up-front is that you remove the ability to avoid the heuristic, and as a heuristic it might turn out to be the wrong thing to do.

If you want a list, why not use `ConfigValueFactory.fromAnyRef(myList)` for example?

derrickburns:
I tend to agree about your concerns for embedding the heuristic. I think the best solution is to add a flag to parseMap for when to use the heuristic.

havocp:
You may need to explain your use case to me. When are you wanting parseMap to do this?

derrickburns:
I have a YAML file reader that outputs maps.

derrickburns:
Actually, the YAML file reader is part of a custom configuration system that  I wrote several years ago. I like yours better l, so I am converting. The only common interface is maps, but because of the ambiguity in the encoding of arrays, I have this problem.

derrickburns:
So the crux of the problem is the ambiguity in the encoding of arrays as keys/paths in maps. Allowing one to specify how to deal with this ambiguity seems reasonable to me.

havocp:
There's no ambiguity - lists are always lists. The only feature that converts is that if you ask for a list with Config.getList and it is an object instead, we will attempt to convert on the fly just as we will convert an integer to a long or what have you. In the actual data structure and file format there's never ambiguity.

Doesn't YAML support lists? I thought it was a JSON superset. But in any case, why not convert to a list in the YAML wrapper code before you parseMap, the conversion code will have a clear explanation for existing if it's in that spot.

derrickburns:
The ambiguity is outside your software. 

JSON cannot be represented as a map from paths to strings unambiguously (unless one changes the encoding of paths to capture the missing bit).

derrickburns:
To give more context...

My system was derived from the system used in Elasticsearch which allows specification of Settings via JSON or YAML or Java properties. These representations cannot be unified into the JSON object model without conflict: Java  properties "x.y" and "x.y.z" are legal but cannot be mapped as is onto the JSON object model. So, the least common denominator is a map from paths to strings. However, this model is ambiguous, since both

 x : [ foo, bar]

And

x: { 0:foo, 1: bar}

Both get  mapped to

x.0 = foo
x.1 = bar


A

derrickburns:
I think that your package should address the ambiguity because you provide a method to construct a Config from a map<path,string>. If you did not provide such s method, then you would not see the ambiguity.

havocp:
It isn't a Map<path, string> though? it's a Map<path, Object> where the Object can be anything that fromAnyRef understands (such as list, string, int, float, boolean, null). You can put a list in there no problem AFAIK.

derrickburns:
The input is a map from paths (e.g. "x.y.z") to strings. This is a flattened representation. There are no objects.

typesafehub-validator:
Hi @derrickburns,

derrickburns:
I modified the pull request so that the new behavior can be selected via a
flag.  I think that this addresses your concerns and mine.

havocp:
Thanks for explaining more. As best I can tell right now I feel like you would probably be the only user of this setInferLists feature. I'm not saying your use case is wrong or bad, but it is specific to having an existing library that generates objects always and never lists. We deal with this situation in Config itself (for the properties parser) in a different way, which is to defer reinterpreting the object until someone calls Config.getList, and for the JSON and HOCON parsers we keep the list around from the start.

One side note, do be careful that "path" means the same thing to you in your library that it does in this library; path is a potentially-escaped expression here that can represent any series of JSON field names, even if they contain spaces and periods. ConfigUtil.splitPath and ConfigUtil.joinPath are available to split and reassemble paths.

I can put together a comparator if you want, to address my suggestion on the other PR, but it might take me some time to get to it. I'm going to schedule a big batch of work on Config sometime in the next month or two I think, to get through everything all at once.

PR #382, opened 2016-03-30
Issue 380 - Empty string as a valid value for empty list

PhilippeCollin:
As stated in #380 by @michellemay (colleague of mine), it is currently impossible to override a config value with an empty list using standard command line arguments (e.g. `-Dfoo=`).

This pull request attempts to solve the problem by allowing empty strings as valid values for lists and by converting them to empty lists.

This means `foo=""` would become valid for all `getList` methods and all would return an empty list of appropriate type.
It would also be possible to override a config value with an empty list using command line arguments like so: `-Dfoo=` .

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java
+++ b/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java
@@ -48,7 +48,12 @@ final class DefaultTransformer {
                 }
                 break;
             case LIST:
-                // can't go STRING to LIST automatically
+                // can't go STRING to LIST automatically unless the string is empty
+                if (s.equals("")) {
+                    // Convert empty string to empty list
+                    ArrayList<AbstractConfigValue> emptyList = new ArrayList<AbstractConfigValue>();
+                    return new SimpleConfigList(value.origin(), emptyList);
+                }
                 break;
             case OBJECT:
                 // can't go STRING to OBJECT automatically
--- a/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java
+++ b/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java
@@ -838,7 +838,8 @@ final class SimpleConfig implements Config, MergeableValue, Serializable {
             }
         } else if (referenceType == ConfigValueType.LIST) {
             // objects may be convertible to lists if they have numeric keys
-            if (value instanceof SimpleConfigList || value instanceof SimpleConfigObject) {
+            // empty strings may also be converted to lists
+            if (value instanceof SimpleConfigList || value instanceof SimpleConfigObject || value instanceof ConfigString) {
                 return true;
             } else {
                 return false;
--- a/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java
+++ b/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java
@@ -838,7 +838,7 @@ final class SimpleConfig implements Config, MergeableValue, Serializable {
             }
         } else if (referenceType == ConfigValueType.LIST) {
             // objects may be convertible to lists if they have numeric keys
-            // empty strings may also be converted to lists
+            // strings may also be convertible to lists if empty
             if (value instanceof SimpleConfigList || value instanceof SimpleConfigObject || value instanceof ConfigString) {
                 return true;
             } else {
--- a/config/src/test/java/beanconfig/ArraysConfig.java
+++ b/config/src/test/java/beanconfig/ArraysConfig.java
@@ -11,6 +11,7 @@ import com.typesafe.config.ConfigValue;
 public class ArraysConfig {
 
     List<Integer> empty;
+    List<Integer> fromEmptyString;
     List<Integer> ofInt;
     List<String> ofString;
     List<Double> ofDouble;
@@ -32,6 +33,10 @@ public class ArraysConfig {
     public void setEmpty(List<Integer> empty) {
         this.empty = empty;
     }
+
+    public List<Integer> getFromEmptyString() { return fromEmptyString; }
+
+    public void setFromEmptyString(List<Integer> fromEmptyString) { this.fromEmptyString = fromEmptyString; }
 
     public List<Integer> getOfInt() {
         return ofInt;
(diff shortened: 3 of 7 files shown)


michellemay:
Good!  Thanks Philippe

config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java
```diff
@@ -48,7 +48,12 @@ static AbstractConfigValue transform(AbstractConfigValue value,
                 }
                 break;
             case LIST:
-                // can't go STRING to LIST automatically
+                // can't go STRING to LIST automatically unless the string is empty
+                if (s.equals("")) {
+                    // Convert empty string to empty list
+                    ArrayList<AbstractConfigValue> emptyList = new ArrayList<AbstractConfigValue>();
```
havocp: `Collections.<AbstractConfigValue> emptyList()` would work here I think, avoiding an allocation

config/src/test/scala/com/typesafe/config/impl/ConfigBeanFactoryTest.scala
```diff
@@ -46,10 +46,9 @@ class ConfigBeanFactoryTest extends TestUtils {
             ConfigBeanFactory.create(config, classOf[ValidationBeanConfig])
         }
 
-        val expecteds = Seq(Missing("propNotListedInConfig", 77, "string"),
-            WrongType("shouldBeInt", 78, "number", "boolean"),
-            WrongType("should-be-boolean", 79, "boolean", "number"),
-            WrongType("should-be-list", 80, "list", "string"))
```
havocp: it isn't clear to me whether this should have been removed - was the test using an empty string? could we keep the test by changing it to a non-empty string? It seems like we'd still want to test this case for non-empty strings ?

havocp:
Thanks for working on this!

I think a helpful test to add would be one that does concatenation or merging between an empty string and a list. Like

```
foo = ""
bar = ${foo} [ 1, 2, 3 ]
```

or perhaps
```
foo = ""
bar += 3
```

I'm not sure it's easy to make this work, but whatever it does we should ensure the test verify what it does.

havocp:
I'm not sure whether/how we handle that for the pretend-a-numeric-indexed-object-is-a-list case, either. But if we do handle it for that, we could do the same here.

PhilippeCollin:
Thank you for your comments! 
I managed to make pretty much everything you said work (concatenation with empty strings for lists and indexed objects). 

For indexed objects, I had to make it so that `foo=""` is also a valid for empty non-indexed objects, which I thought made sense anyway. That way, all objects can also be overridden with command line parameters using the same syntax as for lists. 

There is a few problems I ran into which make me think it may not be realistic, after all, to implement such feature and to have consistent behavior across all existing features.

The way I got it to work normally is, when a config value is parsed, as soon as there is one object or list inside the value, all unquoted whitespace is removed so that it's not joined to empty strings, as it would be, normally, for string values. 

That way, there is no ambiguity in `ConfigConcatenation#join` between :
```
foo=""
bar=${foo} [1, 2] (with whitespace)
```
and
```
foo=" "
bar=${foo}[1, 2] (no whitespace)
```

Here is one problem : 
```
foo=""
bar=""
x=${foo} ${bar} (with unquoted whitespace)
```
Since there is no object or list, I have no way of knowing that I should ignore unquoted whitespace.
Hence, `x` gets a values of `x=" "` (single space).
Then if I call `getList` on that, I obviously get a WrongType exception, and I doubt we would want to accept quoted whitespace as valid list/object values.

There is also some unexpected behavior which may come up such as: 
```
foo=""
bar=[${foo}, [1], [1]]
```
You may want to have `bar=[ [], [1], [1] ]` but we can't really know that.
So, of course you get `bar=[ [""], [1], [1] ]`.

So, I'm no sure we can make this work seamlessly anymore.

PhilippeCollin:
An idea could possibly be to have reserved string symbols for empty arrays/objects. Something like `foo="[]"` and `bar="{}"` and to parse that right upfront as List or Object, but of course these are also valid string values.

havocp:
Thanks for digging in - sometimes the complexities aren't obvious up front!

Idea I just had that could solve this and some similar issues; what about a thing like this:

```
java -Dconfig.literal="foo.bar=[]\nblah={}"
```

Where we take the value of `config.literal`, parse it as a config file (it can be a whole bunch of stuff in there), and insert it in the config stack at the same time we insert system properties (probably properties should override the stuff from `config.literal`, but presumably that case wouldn't arise often).

There might be a better name than `config.literal` ... `config.insert`, `config.extra`, I don't know.

This would not be analogous to `-Dconfig.file` because it wouldn't replace application.conf, it would be in `ConfigFactory.defaultOverrides()` instead.

Using this is a little bit annoying because the quoting rules are double-layered; you have shell quoting followed by JSON quoting. But at least it would make it possible to insert any and all config values on the command line, and even do things like `-Dconfig.literal='foo.bar += 42'` so you could append values to a list for example.

havocp:
maybe config.overrides as the name...

config/src/test/java/beanconfig/ArraysConfig.java
```diff
@@ -11,6 +11,7 @@
 public class ArraysConfig {
 
     List<Integer> empty;
+    List<Integer> fromEmptyString;
```
havocp: Codacy Issue found: [Use explicit scoping instead of the default package private level](https://www.codacy.com/app/hp/config/file/2444096851/issues/source?bid=3245965&fileBranchId=3245971#l14)

config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java
```diff
@@ -48,7 +48,12 @@ static AbstractConfigValue transform(AbstractConfigValue value,
                 }
                 break;
             case LIST:
-                // can't go STRING to LIST automatically
+                // can't go STRING to LIST automatically unless the string is empty
+                if (s.equals("")) {
```
havocp: Codacy Issue found: [Position literals first in String comparisons](https://www.codacy.com/app/hp/config/file/2444096817/issues/source?bid=3245965&fileBranchId=3245971#l52)

config/src/main/java/com/typesafe/config/impl/SimpleConfig.java
```diff
@@ -838,7 +838,8 @@ private static boolean haveCompatibleTypes(ConfigValueType referenceType, Abstra
             }
         } else if (referenceType == ConfigValueType.LIST) {
             // objects may be convertible to lists if they have numeric keys
-            if (value instanceof SimpleConfigList || value instanceof SimpleConfigObject) {
+            // strings may also be convertible to lists if empty
+            if (value instanceof SimpleConfigList || value instanceof SimpleConfigObject || value instanceof ConfigString) {
```
havocp: Codacy Issue found: [Avoid unnecessary if..then..else statements when returning booleans](https://www.codacy.com/app/hp/config/file/2444096835/issues/source?bid=3245965&fileBranchId=3245971#l842)

PR #517, opened 2017-10-23
Some more link fixes after the rename

2m:


ktoso:
Indeed resolves https://github.com/lightbend/config/issues/516 
THanks!

PR #570, opened 2018-07-14
@Optional in getters for properties without field

radist-nt:
Solutions and test for #567.


radist-nt:
I have signed CLA.

havocp:
Thank you!

PR #571, opened 2018-07-15
support for property files with object-string conflicts in path structure

radist-nt:
Proposition to work with problem described in #493.
Didn't managed to setup scala toolchain so unittests are in java.


ennru:

PR #576, opened 2018-07-31
#575 Remove wording about apps having reference.conf

ktoso:
Resolves https://github.com/lightbend/config/issues/575


ktoso:
Thanks for merging :)

PR #602, opened 2018-12-03
added option to switch off validation for ConfigBeanFactory use

dhakehurst:
Useful in its own right I believe, but this also provides a work around for #440

I have beans that I want instantiated by the config file. But I don't want to set every property of the bean. The bean itself sets default values for some properties. I cannot add the '@Optional' annotation to the bean class properties.

switching off the validation allows this.

(I'm using this library for a live project, it is very useful, thanks for all who have contributed towards it.
I need this capability ASAP, and I would rather use an official release over my patched version.)

I have signed the Typesafe Contributor License Agreement online

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@
 // Release tags should follow: http://semver.org/
 import scalariform.formatter.preferences._
 
-ThisBuild / git.baseVersion         := "1.3.0"
+ThisBuild / git.baseVersion         := "1.3.4"
 ThisBuild / organization            := "com.typesafe"
 ThisBuild / Compile / scalacOptions := List("-unchecked", "-deprecation", "-feature")
 ThisBuild / Test / scalacOptions    := List("-unchecked", "-deprecation", "-feature")
--- a/config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
+++ b/config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
@@ -44,6 +44,44 @@ public class ConfigBeanFactory {
      *     Can throw the same exceptions as the getters on <code>Config</code>
      */
     public static <T> T create(Config config, Class<T> clazz) {
-        return ConfigBeanImpl.createInternal(config, clazz);
+        return ConfigBeanImpl.createInternal(config, clazz, true);
+    }
+
+    /**
+     * Creates an instance of a class, initializing its fields from a {@link Config}.
+     *
+     * Example usage:
+     *
+     * <pre>
+     * Config configSource = ConfigFactory.load().getConfig("foo");
+     * FooConfig config = ConfigBeanFactory.create(configSource, FooConfig.class);
+     * </pre>
+     *
+     * The Java class should follow JavaBean conventions. Field types
+     * can be any of the types you can normally get from a {@link
+     * Config}, including <code>java.time.Duration</code> or {@link
+     * ConfigMemorySize}. Fields may also be another JavaBean-style
+     * class.
+     *
+     * Fields are mapped to config by converting the config key to
+     * camel case.  So the key <code>foo-bar</code> becomes JavaBean
+     * setter <code>setFooBar</code>.
+     *
+     * @since 1.3.4
+     *
+     * @param config source of config information
+     * @param clazz class to be instantiated
+     * @param validate passing false means that the config will not be validated against the bean's implied schema 
+     * @param <T> the type of the class to be instantiated
+     * @return an instance of the class populated with data from the config
+     * @throws ConfigException.BadBean
+     *     If something is wrong with the JavaBean
+     * @throws ConfigException.ValidationFailed
+     *     If the config doesn't conform to the bean's implied schema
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -4,7 +4,7 @@
 // Release tags should follow: http://semver.org/
 import scalariform.formatter.preferences._
 
-ThisBuild / git.baseVersion         := "1.3.0"
+ThisBuild / git.baseVersion         := "1.3.4"
```
havocp: This is an unrelated change I guess, ideally not in this PR

config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
```diff
@@ -44,6 +44,44 @@
...
+     *
+     * Fields are mapped to config by converting the config key to
+     * camel case.  So the key <code>foo-bar</code> becomes JavaBean
+     * setter <code>setFooBar</code>.
+     *
+     * @since 1.3.4
+     *
+     * @param config source of config information
+     * @param clazz class to be instantiated
+     * @param validate passing false means that the config will not be validated against the bean's implied schema
```
havocp: I don't love the boolean param ( long story https://blog.ometer.com/2011/01/20/boolean-parameters-are-wrong/ , tldr `create(config, clazz, false)` doesn't read well).

An alternative would be an Options object like ConfigParseOptions / ConfigResolveOptions, or an enum like `ValidationMode { NONE, AUTOMATIC }` or something.

config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
```diff
@@ -44,6 +44,44 @@
...
+     * The Java class should follow JavaBean conventions. Field types
+     * can be any of the types you can normally get from a {@link
+     * Config}, including <code>java.time.Duration</code> or {@link
+     * ConfigMemorySize}. Fields may also be another JavaBean-style
+     * class.
+     *
+     * Fields are mapped to config by converting the config key to
+     * camel case.  So the key <code>foo-bar</code> becomes JavaBean
+     * setter <code>setFooBar</code>.
+     *
```
havocp: I think this kinda needs to explain what happens if we disable validation and it's invalid. I believe the answer is stuff gets left as whatever it was set to originally in the bean (stuff does not get assigned). But full-on type errors (assign an integer to a list or something) maybe still throw?

config/src/test/scala/com/typesafe/config/impl/ParseableReaderTest.scala
```diff
@@ -2,29 +2,29 @@ package com.typesafe.config.impl
...
-      ConfigFactory.parseReader(configInput)
+    @Test
+    def parseIncorrectFormat(): Unit = {
+        val filename = "/test01.properties"
+        val configInput = new InputStreamReader(getClass.getResourceAsStream(filename))
+        val e = intercept[ConfigException.Parse] {
+            ConfigFactory.parseReader(configInput)
+        }
+        assertThat(e.getMessage, containsString("Expecting end of input or a comma, got '^'"))
     }
```
havocp: are there changes in this file or just accidental reformatting?

dhakehurst: I changed nothing here.

build.sbt
dhakehurst: the current release is 1.3.3, so I don't know why this was set to 1.3.0.
I changed it to 1.3.4, so that my local publication is at a higher version than the release.
(perhaps it should still read 1.3.3 or 1.3.0?)

config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
dhakehurst: That would be an acceptable alternative. though I am fine with booleans.

kringol: I am also looking forward to this feature, since i also need it. 
But i would also think it would be nicer to have more control on the validation. For example to keep the "type" validation for non-null fields, but ignore the optional/null value check, thus allowing null values but validating the type on non-null ones. And similar. 
Currently the boolean is disabling/enabling everything

kringol:
Just added a PR to your own fork with some small changes to match the mechanism of other parts of the library and enable future settings if needed, plus a mini test
https://github.com/dhakehurst/config/pull/1

Thanks :)

> Feel free to add to this pull request or superceed it. I have done all i am able to do at this time.

avillev:
Hello there !! @havocp @dhakehurst 
Do we have new about that change?
Does someone plan to have it merged into the project?

dhakehurst:
I do not currently have more time to work on this

kringol:
I will continue this change on a new PR instead then. 
I really want this moving forward and been waiting for it

-> https://github.com/lightbend/config/pull/616

havocp:
Closing this since it moved to #616

PR #616, opened 2019-02-19
Bean validation config to allow missing values or other settings

kringol:
_Continuation of https://github.com/lightbend/config/pull/602. Taking into account the feedback, etc_

The purpose of this change is to provide a mechanism to configure the bean factory internal workings, and in this particular case, to provide a setting to allow missing field values and keep whatever value was originally assigned to the bean or null, instead of throwing an a validation error.
By default such config is disabled, to keep the current behavior as it is and provide a configuration mechanism similar to other sections of the library

### DIFF
--- a/config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
+++ b/config/src/main/java/com/typesafe/config/ConfigBeanFactory.java
@@ -44,6 +44,44 @@ public class ConfigBeanFactory {
      *     Can throw the same exceptions as the getters on <code>Config</code>
      */
     public static <T> T create(Config config, Class<T> clazz) {
-        return ConfigBeanImpl.createInternal(config, clazz);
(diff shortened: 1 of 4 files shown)


rkhmelichek:
Are there plans to merge this pull request?
I'm also in need for a solution to #440

kringol:
Waiting for @havocp and the team to review/approve prepare the next release i guess. Not sure about the timelines.

avillev:
@havocp @andreaTP @johanandren @2m 

Sorry to tag that many mergers but, what should we do to get feedback or merge that PR? It has been a while and I kind of need that improvement too for my project 👍

tadhgpearson:
This functionality would be really useful (for me anyway!) but this PR's been open for a long time. What is the barrier to merging this?

ennru:

PR #632, opened 2019-05-07
Test on JDK 11

raboof:


.travis.yml
```diff
@@ -7,11 +7,18 @@ before_install:
   # using jabba for custom jdk management
   - curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh | bash && . ~/.jabba/jabba.sh
```
marcospereira:
```suggestion
   - curl -sL https://git.io/jabba | bash && . ~/.jabba/jabba.sh
```

marcospereira: So that the build will always use an updated version of Jabba.

.travis.yml
```diff
@@ -5,13 +5,20 @@ language: scala
 
 before_install:
   # using jabba for custom jdk management
-  - curl -sL https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh | bash && . ~/.jabba/jabba.sh
-  - jabba install adopt@1.8.202-08
-  - java -version
+  - curl -sL https://git.io/jabba | bash && . ~/.jabba/jabba.sh
+  - jabba install adopt@1.8.212-03
+  - jabba install adopt@1.11.0-3
```
ennru: You could possibly use the `~` notation to use the latest build of the JDK versions.

marcospereira: Not working right now for `adopt@~1.11`. I assume jabba is failing to correct check versions such as `adopt@1.11.0-2` because of the `-2` section. Anyway, should not block this PR.

PR #642, opened 2019-07-08
Fix compiler warnings in tests

naferx:
Aside from this PR, `Config#getMilliseconds()` and `Config#getNanoseconds()` have been marked as deprecated for a while. Makes sense to remove them for next minor release.


ennru:

PR #660, opened 2019-12-08
Inherit optional and getter evaluation

### LINKED ISSUE #659: Inheritable @Optional and evaluation at getter even if field exists
I'm experimenting with AWS CDK in Java and have to build some Property beans for cloud constructs. Usually it is a mix of mostly predefined options  and some options that are later resolved (e.g. IDs of dependent constructs). Therefore I thought it might be worth a try to mix immutable.io with lightbend config and except of one small issue everything works out fine.
When I annotate a property in an immutable bean class with @Optional the generated bean class doesn't have the annotation because @Optional isn't marked as inheritable. In my mind it should be and I haven't found a reason in the lightbend config discussions why not. (Maybe to harsh but in my mind whenever you override a field in an inherited bean and don't want it to inherit the parent annotations, you already have a code smell.)
the second point belongs to #570 where the @Optional at getters is only evaluated if the field does not exist. Even if the field exists and doesn't have the annotation, the corresponding getter should also be evaluated for the annotation.
I will try to deliver a PR as a discussion base.

derKrischan:
Added @Inherited to @Optional and an extra evaluation for the presence of @Optional at getter if the field exists. All tests passing but maybe someone can give me a hint where I can implement an extra test for my changes if necessary.
I signed the Lightbend CLA. 
closes #659


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
Hi @fabiofumarola,

derKrischan:
Tried to rewrite git history for my username and mail to match the CLA but that went wrong :-) . I'll give it another try.

PR #661, opened 2019-12-08
Inherit optional and getter evaluation

### LINKED ISSUE #659: Inheritable @Optional and evaluation at getter even if field exists

derKrischan:
Added @inherited to @optional and an extra evaluation for the presence of @optional at getter if the field exists. All tests passing but maybe someone can give me a hint where I can implement an extra test for my changes if necessary.
I signed the Lightbend CLA.
closes #659

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
+++ b/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
@@ -284,11 +284,11 @@ public class ConfigBeanImpl {
     }
 
     private static boolean isOptionalProperty(Class beanClass, PropertyDescriptor beanProp) {
+        if (beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0) {
+            return true;
+        }
         Field field = getField(beanClass, beanProp.getName());
-        return field != null
-                ? beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0
-                        || field.getAnnotationsByType(Optional.class).length > 0
-                : beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0;
+        return field != null && field.getAnnotationsByType(Optional.class).length > 0;
     }
 
     private static Field getField(Class beanClass, String fieldName) {
--- a/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
+++ b/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
@@ -285,7 +285,10 @@ public class ConfigBeanImpl {
 
     private static boolean isOptionalProperty(Class beanClass, PropertyDescriptor beanProp) {
         Field field = getField(beanClass, beanProp.getName());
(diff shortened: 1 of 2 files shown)


config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
```diff
@@ -285,7 +285,10 @@ private static boolean hasAtLeastOneBeanProperty(Class<?> clazz) {
 
     private static boolean isOptionalProperty(Class beanClass, PropertyDescriptor beanProp) {
         Field field = getField(beanClass, beanProp.getName());
-        return field != null ? field.getAnnotationsByType(Optional.class).length > 0 : beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0;
+        return field != null
+                ? field.getAnnotationsByType(Optional.class).length > 0
+                        || beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0
```
radist-nt, line 290: Maybe 
`beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0` should be checked before calling getField ?
Method getField modifies field via reflection and optionally recursively calls itself, so it could be costly.

config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
```diff
@@ -286,8 +286,8 @@ private static boolean hasAtLeastOneBeanProperty(Class<?> clazz) {
     private static boolean isOptionalProperty(Class beanClass, PropertyDescriptor beanProp) {
         Field field = getField(beanClass, beanProp.getName());
         return field != null
-                ? field.getAnnotationsByType(Optional.class).length > 0
-                        || beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0
+                ? beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0
```
radist-nt, line 289: I meant, `beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0` could be evaluated at the beginning of method:
```
if (beanProp.getReadMethod().getAnnotationsByType(Optional.class).length > 0)
   return true;
Field field = getField(...);
return field != null && ...

ennru:

PR #711, opened 2020-10-02
Improve set array values section

andreaTP:


ignasi35 (approved):
Brilliant. That's a lot clearer and more detailed. Thanks @andreaTP

havocp (approved):

Issue #145, opened 2014-02-16
Support for masking values when rendering


gregsymons:
In my app, I'm exposing the combined configuration on a REST endpoint. For the most part, `configuration.root.render()` is good enough. However, I'd like to mask out sensitive keys that hold things like passwords. I've been able to implement masking like this:

``` scala
val shouldMask = "(secret|password)".r.unanchored
def masked(value: ConfigValue) = {
   ConfigValueFactory.fromAnyRef("MASKED", "Value masked for security reasons")
}

def safeConfiguration(configuration: Config) = {
  configuration.entrySet.asScala.foldLeft(configuration) { 
    case (config, (key@shouldMask(blacklisted), value)) => {
      config.withValue(key, masked(value))
    }
    case (config, (key, _)) => {
      config
    }
  }
} 
```

The problem is that when I construct the new `ConfigValue`, I lose the original origin information. Ideally, I'd like to just add the `"Value masked for security reasons"` to the end of the existing comments, if any. However, there's no way to construct a new `ConfigOrigin` from the original origin, nor a way to pass one in if I could.

Really, this masking thing could be an additional `ConfigRenderOptions` option and hidden in `render`. Maybe you could do it like this:

``` scala
def safeConfiguration(configuration: Config) = {
  val shouldMask = "(secret|password)".r.unanchored
  configuration.render(ConfigRenderOptions.defaults.withFilter {
    case (shouldMask(_), _) => true
    case _ => false
  }
```

Though that may be too Scala an interface for this library :smile:. An easier solution would be to just expose a way to "mutate" a ConfigOrigin and pass it into `ConfigValueFactory.fromAnyRef` so I can implement it myself without either making the "pretty" output of `render` ugly or losing the origin information. Or suggest a way I can do it with the existing API:)

shanielh:
@havocp - Can we set another way to quote values (i.e, `p"myPassword"` for passwords), I've noticed that when rendering the configuraiton - We're able to see if the value is quoted or not.

viktorklang:
My recommendation: don't put passwords in config files and don't rely on config masking to "secure" anything.

clehene:
@viktorklang so use config for everything but for secrets use something else. Then why not use that something else for everything?

This question is not about storing secrets in a config *file*, but rather having it passed through a config object. As Typesafe config handles env variables integration (yet still displays them freely) it would be the normal mechanism to get the secrets in, provided it would offer the masking capability.

viktorklang:
[...]

PR #724, opened 2021-01-27
Support naive masking of config values

norrs:
You can supply a regex pattern to match against Config key names that it
will mask it corresponding config value.  If the config value is a
config reference, it will keep the config reference.

This is a naive approach, and secrets will not be masked if the config
reference key is also not matching the maskRegex.

This is useful for printing configuration "safely" (read: best effort)
without exposing secrets.

First version for feature:
https://github.com/lightbend/config/issues/145

### DIFF
--- a/config/src/test/scala/com/typesafe/config/impl/ConfigTest.scala
+++ b/config/src/test/scala/com/typesafe/config/impl/ConfigTest.scala
@@ -1149,17 +1149,20 @@ class ConfigTest extends TestUtils {
     @Test
     def renderRoundTrip() {
         val allBooleans = true :: false :: Nil
+        val noMasking = Nil
         val optionsCombos = {
             for (
                 formatted <- allBooleans;
                 originComments <- allBooleans;
                 comments <- allBooleans;
-                json <- allBooleans
+                json <- allBooleans;
+                maskings <- noMasking
             ) yield ConfigRenderOptions.defaults()
                 .setFormatted(formatted)
                 .setOriginComments(originComments)
                 .setComments(comments)
                 .setJson(json)
+                .setMaskingRegex(maskings)
         }
 
         for (i <- 1 to 10) {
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @norrs,

norrs:
This comes rather late (only ~7 years) according to #145 , but would this be good enough for first iteration? 
@gregsymons @havocp

config/src/test/scala/com/typesafe/config/impl/ConfigTest.scala
```diff
@@ -1149,17 +1149,20 @@ class ConfigTest extends TestUtils {
     @Test
     def renderRoundTrip() {
         val allBooleans = true :: false :: Nil
+        val noMasking = Nil
```
norrs, line 1152: Supplied only Nil here for the roundtrip test. Unsure how to elegantly assert correctly when masking is enabled.

eneveu:
It would be nice if we could plug a sanitizer that takes a key/value and returns the masked value.

This would let us use something similar to the Spring Boot Sanitizer:
https://github.com/spring-projects/spring-boot/blob/v2.5.0-M3/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java

I really like that it can also mask passwords inside URLs, like `http://user:******@localhost:8080` or `amqp://username:******@host/`

norrs:
@eneveu : Shouldn't be harder then replicating the logic of the following lines:

```
if (keyIsUriWithUserInfo(pattern)) {
  return sanitizeUris(value.toString());
}
return "******";
```

But I'd first like to see if upstream even cares about this feature or not, and if they do, we can follow up with a version 2. The joy of software development, we can always improve step by step ;-)

eneveu:
What I meant is that, instead of an API that exposes a specific implementation detail (a masking Regex), it would be cool if the API allowed the end-user to customise the way values are sanitized.

So, instead of passing a "Pattern maskRegex" to the "ConfigRenderOptions", we could pass a "ValueRenderer":

```
trait ValueRenderer {
  def render(key: String, value: ConfigValue): ConfigValue
}
```

And then provide a simple ValueRenderer that masks values when the key matches a regex. And then end-users can pass their own ValueRenderer with more complex masking behaviour.

Maybe "ValueRenderer" is not the best name. It could be "ValuePrinter" or "ValueSanitizer" if we want to make it explicit that the goal is to sanitize values.

Also, the "key" should probably be the full path, to let people match against the full path in their regex.

eneveu:
Another option would be to add a way to create a `ConfigValue` while passing the origin and comments of a previous value.
If we do this, we can create an external `Sanitizer` that walks the Config tree and sanitizes values as needed.
More info here: https://github.com/lightbend/config/issues/145#issuecomment-827059275

ennru:

PR #758, opened 2022-02-01
Avoid instantiating arbitrary classes

raboof:
This PR avoids the use of `Class#newInstance`, which is
deprecated in Java 9.

In particular, previously you could set the `config.strategy`
system to an arbitrary class that would get instantiated even
if it was not a subclass of `ConfigLoadingStrategy`. This is
now checked before instantiating the class.

The previous behavior could arguably be considered a security
concern when an attacker has write access to system properties,
though in such a scenario there are likely many other ways to
load arbitrary code.


havocp (approved):

config/src/main/java/com/typesafe/config/ConfigFactory.java
```diff
@@ -1241,7 +1242,11 @@ private static ConfigLoadingStrategy getConfigLoadingStrategy() {
 
         if (className != null) {
             try {
-                return ConfigLoadingStrategy.class.cast(Class.forName(className).newInstance());
+                return Class.forName(className).asSubclass(ConfigLoadingStrategy.class).getDeclaredConstructor().newInstance();
+            } catch (InvocationTargetException e) {
+                Throwable cause = e.getCause();
+                if (cause == null) throw new ConfigException.BugOrBroken("Failed to load strategy: " + className, e);
+                else throw new ConfigException.BugOrBroken("Failed to load strategy: " + className, cause);
```
vlsi, line 1249: ❤️  unwrapping `InvocationTargetException` is great

PR #776, opened 2022-11-09
Add ConfigFloat class to wrap java.lang.Float.

daniel-jasinski:
Instances of java.lang.Float were previously wrapped using ConfigDouble. This had the unfortunate side effect of adding junk values when converting to double. For example float value `1.64e-7f` would be rendered as `1.6399999935856613E-7`.

This issue was discovered when testing the HOCON module of the Kotlin Serialization library.


ennru:

PR #780, opened 2023-01-05
CI on Java 17; GH action versions

ennru:
- Add Java 17 to the build matrix
- Use current major versions of the official GH actions


ennru:
I guess the sbt upgrade from https://github.com/lightbend/config/pull/756 is necessary to test on Java 17.

ennru:
Updated #756  instead.

Issue #796, opened 2023-08-09
Cannot provide environment variables programmatically


przemek-pokrywka:
Many container-era applications follow the [12-factor-app specification](http://12factor.net) which puts focus on environment variables. The environment variable substitutions used in the configuration files can not be easily tested, which is a problem when the configuration is complex and when one wants to run multiple parallel tests, each with different variables substituted.

Why is substituting environment variables programmatically difficult: the JVM does not expose any API for altering an environment variable unless you resort to some [reflection API hacks](https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java). Also, even then, the variable gets substituted globally, which prevents you from running different similar tests in parallel.

There exist libraries that address these issues by letting the user programmatically alter the environment variables in specified scope (ZIO Config, thanks to ZIO [TestSystem](https://zio.dev/reference/test/services/system/)), but when you integrate them with this library, they fall short, because Lightbend Config only ever uses the JVM's System class directly.

Ideally, Lightbend Config would expose an API that would let one substitute environment variables programmatically, for the time of loading/resolving of configuration. Such an API could then be used directly or through testing libraries that support the notion of env. var. substitutions.

PR #797, opened 2023-08-09
Enable programmatic overriding of environment variables

przemek-pokrywka:
This new API allows users to easily test how environment variables substitutions would work in their complex configurations, with:
- no need for [ugly reflection hacks](https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java) (JVM provides no way to alter env. variables)
- and without being forced to run their tests sequentially (as even with the hack, the env. variables are global).

Besides environment variables, the new API also lets one substitute system properties and standard error stream, everything in a scoped way (nothing is set globally, only per caller thread).

Caveat: the substitutions work only in the current thread, however, this should fit 99.9% of users, since it is rather unheard of, that one does multi-threaded computations during the configuration loading/resolving process.

References #796

### DIFF
--- a/config/src/main/java/com/typesafe/config/ConfigFactory.java
+++ b/config/src/main/java/com/typesafe/config/ConfigFactory.java
@@ -1137,13 +1137,13 @@ public final class ConfigFactory {
 
         // override application.conf with config.file, config.resource,
         // config.url if requested.
-        String resource = System.getProperty("config.resource");
+        String resource = SystemOverride.getProperty("config.resource");
         if (resource != null)
             specified += 1;
-        String file = System.getProperty("config.file");
+        String file = SystemOverride.getProperty("config.file");
         if (file != null)
             specified += 1;
-        String url = System.getProperty("config.url");
+        String url = SystemOverride.getProperty("config.url");
         if (url != null)
             specified += 1;
 
@@ -1238,7 +1238,7 @@ public final class ConfigFactory {
     }
 
     private static ConfigLoadingStrategy getConfigLoadingStrategy() {
-        String className = System.getProperties().getProperty(STRATEGY_PROPERTY_NAME);
+        String className = SystemOverride.getProperties().getProperty(STRATEGY_PROPERTY_NAME);
 
         if (className != null) {
             try {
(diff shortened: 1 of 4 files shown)


ThijsBroersen:
+1

mkurz:
This PR likely will never be merged.

https://github.com/lightbend/config#maintained-by
> The "Typesafe Config" library is an important foundation to how Akka and other JVM libraries manage configuration. We at [Lightbend](https://lightbend.com/) consider the functionality of this library as feature complete. We will make sure "Typesafe Config" keeps up with future JVM versions, but will rarely make any other changes.

przemek-pokrywka:
Thanks for chipping in @mkurz, unfortunately, I am aware of that, yet I am trying to make people realize the utility of the change, especially since it is small and pretty well-contained.

Dear @lightbend maintainers, please consider merging this change, as no good workaround for the issue exists.
Why: if one wanted to temporarily replace the JVM's private copy of environment variables through reflection, it will always be exposed to the risk of 1) future JVMs changing their internals *and* 2) concurrent access to the environment from other threads

yantonov:
Even if it's a direct fix for the existing functionality new patches are not applied.

https://github.com/lightbend/config#maintained-by

It was discussed previously, and the conclusion is always the same.

przemek-pokrywka:
> I understand your wish to improve testability of your code

Dear @ennru, dear @lightbend maintainers, not only of my code but also of code maintained by teams, to which the people interested in that issue belong. And, by virtue of transitive dependencies, of code that uses zio-config and similar libraries.
So if your decision was based on the affected auditorium, I would really expect that it's wider than just me, and I would encourage you to reconsider it.

Modern standards of testability are much higher than they were at the time of the conception of this library. You, as the steward, have a unique opportunity to keep the library relevant.

yantonov:
Here https://github.com/lightbend/config/pull/798 extended functionality was merged, although 
a) it's explicitly stated that the library is feature-complete (https://github.com/lightbend/config#maintained-by);
b) but bugs for existing functionality and explicitly stated and therefore expected behavior are not fixed for years.
Great!

dobrynya:
+1

przemek-pokrywka:
Dear @ennru, dear https://github.com/lightbend maintainers,

As correctly pointed out by @yantonov at https://github.com/lightbend/config/pull/797#issuecomment-1778614070, feature completeness should not mean, that the users should cope with the existing problems until they move to an alternative.
Weak testability *is* a problem for anyone, who does non-trivial (or production-critical) things with configuration, and you know very well, that the library gives one great power, linked to great risk sometimes.

As the responsible stewards of the library, which *still* enjoys great popularity within the ecosystem, please do consider merging these changes.

IRus:
Workaround: set env properties into system properties: https://github.com/Heapy/komok/blob/main/komok-tech-config/src/main/kotlin/io/heapy/komok/tech/config/ConfigurationModule.kt#L33

PR #817, opened 2025-06-09
fix: Incorrect abs of hashes in BadMap

johanandren:
Can lead to negative array indexes and throw out of bounds exception on config load

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/BadMap.java
+++ b/config/src/main/java/com/typesafe/config/impl/BadMap.java
@@ -74,7 +74,7 @@ final class BadMap<K,V> {
     }
 
     private static void store(Entry[] entries, Entry e) {
-        int i = e.hash % entries.length;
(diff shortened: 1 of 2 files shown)


johanandren:
Reported over in one of the Pekko issue trackers: https://github.com/apache/pekko/issues/1886

config/src/main/java/com/typesafe/config/impl/BadMap.java
```diff
@@ -74,7 +74,7 @@ private static <K,V> void store(Entry[] entries, int hash, K k, V v) {
     }
 
     private static void store(Entry[] entries, Entry e) {
-        int i = e.hash % entries.length;
```
leviramsey, line 77: I think e.hash here is already abs'd (see line 65)...

johanandren: I hit an exception when trying it out in the test.

Issue #838, opened 2026-04-13
Substitution is evaluated even when hidden by a value that could not be merged with it


myjniak:
According to https://github.com/lightbend/config/blob/v1.4.6/HOCON.md#examples-of-self-referential-substitutions :

```text
If a substitution is hidden by a value that could not be merged with it (by a non-object value) then it is never evaluated and no error will be reported. So for example:

foo : ${does-not-exist}
foo : 42
In this case, no matter what ${does-not-exist} resolves to, we know foo is 42, so ${does-not-exist} is never evaluated and there is no error. 
```

```
p: {"a" : ${y}}
p: {"a" : 42}
p: {"a" : { "x" : 1 }}
```
gets correctly evaluated to {"p":{"a":{"x":1}}}

HOWEVER, if you hide the middle value under a substitution:
```
p: {"a" : ${y}}
p: ${x}
p: {"a" : { "x" : 1 }}
x: {"a" : 42}
```
We get an error:
`Could not resolve substitution to a value: ${y}`

ennru:
Fixed with
- #839

PR #839, opened 2026-04-14
fix: Don't evaluate substitutions hidden by values from resolved objects

johanandren:
References
- #838

Resolving a `ConfigDelayedMergeObject` eagerly resolved every stack entry, evaluating substitutions that the spec says should be skipped when hidden by a value that cannot merge with them. This fix prunes keys from lower-priority object entries whose accumulated merged value already ignores fallbacks, so those substitutions are discarded without evaluation, matching the equivalent direct-literal case.


ennru (approved):
LGTM.

PR #840, opened 2026-04-14
Fix spelling

jsoref:
I have read [Maintained by](https://github.com/lightbend/config?tab=readme-ov-file#maintained-by)

I'm happy to split or drop things if people are interested.

Akka replaced Lightbend, so the CLA link redirects which means you should apply 5808a5deee1d1adea31de96a2a241b4778063ee6 or something very close to it.

The rest are more typical spelling / grammar fixes -- I started on this repository because we use it transitively and I generally send PR contributions to things I/we use. I'm not attached to the corrections, but I find it easier to work w/ code when it doesn't have typos.

PR #843, opened 2026-04-15
CI fixes

jsoref:
In any event, you're going to want to apply 18959b16c41b2fb47ab445183ff9795058d861fd or something like it as the workflows are using obsolete action versions which trigger warnings.

Hopefully, eventually, `apt-key` will disappear, and you'd want to apply fc4bf1388c9a4d43f8f88618f0834ee2f77cf179 or something like it.

References
- https://github.com/lightbend/config/pull/840#issuecomment-4249927939

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,9 @@ jobs:
           - java: 11
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM

PR #846, opened 2026-05-04
fix: BugOrBroken - partially-shadowed object has nested delayed merge

johanandren:
The 1.4.7 fix for #838 (commit 833bacb) added per-key pruning that built a new `SimpleConfigObject` containing only the unshadowed keys and substituted it for end before resolving. The outer `ConfigDelayedMergeObject` still referenced the original (unpruned) end in its stack, so any inner `ConfigDelayedMerge` whose resolution walked the parent chain back up failed `replaceChildInList` with:

> BugOrBroken: tried to replace SimpleConfigObject({...}) which is not in [...]

This fixes that by narrowing the pruning to a whole-entry skip: only continue when every key in end is shadowed by an ignoresFallbacks value in merged. Never substitute a partial copy of end, that changes identity and breaks the parent-chain walk.


patriknw (approved):
LGTM

PR #851, opened 2026-06-02
feat: opt-in unknown key validation for ConfigBeanFactory

### LINKED ISSUE #842: ConfigBeanFactory silently ignores unknown config keys
`ConfigBeanFactory.create()` appears to silently ignore config keys that do not map to any JavaBean property.
 
 I could not find this behavior documented in the README, and the wording around validating against the bean's implied schema made me expect stricter validation.

From the README ([here](https://github.com/lightbend/config/blob/main/README.md#configbeanfactory))
>Creating a bean from a Config automatically validates that the config matches the bean's implied schema. Bean fields can be primitive types, typed lists such as List<Integer>, java.time.Duration, ConfigMemorySize, or even a raw Config, ConfigObject, or ConfigValue (if you'd like to deal with a particular value manually).
 
 ## Observed Behavior
 
 Given a bean like:
 
 ```java
 public class EndpointConfig {
     @Optional
     private String protocol;
 
     public String getProtocol() {
         return protocol;
     }
 
     public void setProtocol(String protocol) {
         this.protocol = protocol;
     }
 }
```

And config with a typo in the key:

**HOCON:**

```hocon
protocl = "https"
```

**JSON:**

```json
{
 "protocl": "https"
}
```

this code does not throw:

```java
 Config config = ConfigFactory.parseString("protocl = \"https\"").resolve();
 // or JSON:
 // Config config = ConfigFactory.parseString("{\"protocl\":\"https\"}",
 //     ConfigParseOptions.defaults().setSyntax(ConfigSyntax.JSON)).resolve();
 
 EndpointConfig bean = ConfigBeanFactory.create(config, EndpointConfig.class);
 System.out.println(bean.getProtocol()); // null
```

Test case:

```java
    @Test
    void misspelledKey_isSilentlyIgnored_fieldRemainsNull() {
        var config = ConfigFactory.parseString("""
                protocl = "https"
                """); // typo: "protocl" instead of "protocol"

        var bean = ConfigBeanFactory.create(config, EndpointConfig.class); // does not throw exception


        assertNull(bean.getProtocol());
    }
```

The unknown key is ignored and the bean property remains unset.

## Clarification / Enhancement Suggestion

At minimum, it would help to document this explicitly in the README / API docs.

If this is intentional, it may also be useful to offer an opt-in strict mode so that unknown config keys can fail validation instead of being silently ignored.

For example, something conceptually like:

 - current default behavior: ignore unknown keys
 - Optional strict behavior: fail if config contains keys that do not map to bean properties

Typos in config keys are easy to make and hard to detect when they are silently ignored. This is especially surprising because missing required bean properties and wrong value types do fail validation.

From reading the implementation, ConfigBeanImpl validates known bean properties but does not appear to check for leftover/unmapped keys after binding. I also did not find an existing setting to enable strict unknown-key validation. If the current behavior is intentional, documenting it would already help a lot.

---

I am happy to contribute, if you think an opt-in strict mode would be acceptable.

mattheworford:
References
- Closes #842

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
+++ b/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
@@ -85,12 +85,12 @@ public class ConfigBeanImpl {
             // find every issue, but it should find common ones).
             List<ConfigException.ValidationProblem> problems = new ArrayList<ConfigException.ValidationProblem>();
             if (!allowUnknownConfigKeys) {
-                for (Map.Entry<String, String> originalName : originalNames.entrySet()) {
-                    String camelName = originalName.getKey();
+                for (Map.Entry<String, String> nameEntry : originalNames.entrySet()) {
+                    String camelName = nameEntry.getKey();
                     if (!beanPropNames.contains(camelName)) {
                         AbstractConfigValue configValue = configProps.get(camelName);
                         problems.add(new ConfigException.ValidationProblem(
-                                Path.newKey(originalName.getValue()).render(),
+                                Path.newKey(nameEntry.getValue()).render(),
                                 configValue.origin(),
                                 "Unknown config setting"));
                     }
--- a/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
+++ b/config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
@@ -40,7 +40,7 @@ public class ConfigBeanImpl {
      * @param clazz class of the bean
      * @return the bean instance
      */
-    public static <T> T createInternal(Config config, Class<T> clazz) {
+    public static <T> T createInternal(Config config, Class<T> clazz, boolean allowUnknownConfigKeys) {
         if (((SimpleConfig)config).root().resolveStatus() != ResolveStatus.RESOLVED)
             throw new ConfigException.NotResolved(
                     "need to Config#resolve() a config before using it to initialize a bean, see the API docs for Config#resolve()");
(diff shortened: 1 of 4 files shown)


config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
```diff
@@ -76,10 +76,26 @@ public static <T> T createInternal(Config config, Class<T> clazz) {
...
+            Set<String> beanPropNames = new HashSet<String>();
+            for (PropertyDescriptor beanProp : beanProps) {
+                beanPropNames.add(beanProp.getName());
+            }
 
             // Try to throw all validation issues at once (this does not comprehensively
             // find every issue, but it should find common ones).
             List<ConfigException.ValidationProblem> problems = new ArrayList<ConfigException.ValidationProblem>();
+            if (!allowUnknownConfigKeys) {
+                for (Map.Entry<String, String> originalName : originalNames.entrySet()) {
```
johanandren: `originalName` reads a bit strange when it is actually a `Entry(camelCase, originalName)`, just `entry` or `nameEntry` would be more self explanatory.

johanandren (commented):
Looks good with one naming nitpick

config/src/main/java/com/typesafe/config/impl/ConfigBeanImpl.java
mattheworford: good call - updated in https://github.com/lightbend/config/pull/851/commits/fc7d977aec45f389bf243fdac3f23d5a0d89cd02

johanandren (approved):
LGTM, thanks!

PR #853, opened 2026-06-03
fix: wrong line number for objects after multiline string

### LINKED ISSUE #852: lineNumber() returns totally wrong line for elements of ConfigObject
I really hope that someone can help with this rather heavy bug in ConfigOrigin.lineNumber().

I order steps in a processing object according to their line numbers. I know all about the discussion about that ConfigObject properties in inherently not ordered - I understand and agree 100%. But in many scenarios where a config is purely defined by end-users, not touched at all by any reference.conf or other resolution strategies, is seems perfectly valid to ask for the origin (specifically the line number) and sort elements by that.

So, I have this config:
```
transformations {
  trim-xml-leading-whitespace = """
    stringTransformation {
        println("Original message: '$it'")
        it.trimStart { it.isWhitespace() || it == '\uFEFF' }.also { println("Trimmed leading whitespace: '${it}'") }
    }
  """
}

pipelines {
  my-pipeline {
    from {
      type = GENERATOR
      count = 1
      message = "foo"
    }
    error-strategy = SHUTDOWN
    processing {
      transformation = trim-xml-leading-whitespace
      xml-to-json = { attribute-prefix = "@" }
    }
    to {
      type = LOGGER
    }
  }
}
``` 

The processing object is the ConfigObject I am focusing on, getting all its root() keys and sorting them by lineNumber.

But in this specific case, it states:
transformation -> lineNumber 19 (perfectly ok!)
xml-to-json -> lineNumber 15 (completely wrong!)

The ConfigOrigin object have identical other elements like originType FILE, the description is identical (they come from the same file name after all) - but WHY on earth is lineNumber for the key xml-to-json resolved to 15??

On line 15 in this example we have a totally unrelated object:
```
      message = "foo"
```

This is *nothing* to do with the processing object! Zero. Nothing.

When I try to write test cases for this specific example, I fail. When my config comes from a pure String, transformed into a Config by ConfigFactory, everything works just fine.

Please help!

mattheworford:
References
- Closes #852

### DIFF
--- a/config/src/test/scala/com/typesafe/config/impl/ConfParserTest.scala
+++ b/config/src/test/scala/com/typesafe/config/impl/ConfParserTest.scala
@@ -917,6 +917,7 @@ class ConfParserTest extends TestUtils {
             "    processing {",
             "      transformation = trim-xml-leading-whitespace",
             "      xml-to-json = { attribute-prefix = \"@\" }",
+            "      xml-to-json-list = [ { attribute-prefix = \"@\" } ]",
             "    }",
             "    to {",
             "      type = LOGGER",
@@ -927,6 +928,7 @@ class ConfParserTest extends TestUtils {
 
         assertEquals(19, conf.getValue("pipelines.my-pipeline.processing.transformation").origin().lineNumber())
         assertEquals(20, conf.getObject("pipelines.my-pipeline.processing.xml-to-json").origin().lineNumber())
+        assertEquals(21, conf.getList("pipelines.my-pipeline.processing.xml-to-json-list").origin().lineNumber())
     }
 
     @Test
--- a/config/src/test/scala/com/typesafe/config/impl/ConfParserTest.scala
+++ b/config/src/test/scala/com/typesafe/config/impl/ConfParserTest.scala
@@ -894,6 +894,42 @@ class ConfParserTest extends TestUtils {
     }
 
     @Test
+    def valuesAfterMultilineStringHaveCorrectOriginLine() {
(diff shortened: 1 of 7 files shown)


Zordid:
Would be *wonderful* if this fix could be released asap... Thanks so much, @mattheworford !

config/src/test/scala/com/typesafe/config/impl/ConfParserTest.scala
```diff
@@ -893,6 +893,42 @@ class ConfParserTest extends TestUtils {
...
+            "    to {",
+            "      type = LOGGER",
+            "    }",
+            "  }",
+            "}"
+        ).mkString("\n"))
+
+        assertEquals(19, conf.getValue("pipelines.my-pipeline.processing.transformation").origin().lineNumber())
+        assertEquals(20, conf.getObject("pipelines.my-pipeline.processing.xml-to-json").origin().lineNumber())
+    }
```
johanandren, line 932: Since we are touching `parseArray` as well, add a test covering that.

Zordid: So the problem I identified also affects arrays? Interesting.

mattheworford: @johanandren added/passes in https://github.com/lightbend/config/pull/853/commits/68ff13390bc580206bb54c2a988f944a409491c7

johanandren (approved):
LGTM, thanks @mattheworford !

PR #856, opened 2026-08-13
Fix resolveWith() for overloaded keys with delayed merge

### LINKED ISSUE #855: Variable substitution fails unexpectedly with overloaded values using resolveWith()
An unexpected error occurs with the following code:

```java
// This FAILS!
Config conf = ConfigFactory.parseString("{ \"one\": \"first\", \"one\": ${variable}}")
        .resolveWith(ConfigFactory.parseString("{ \"variable\": \"second\"}"));
```

The expected behavior is that the value of "one" will be ${variable}, which will resolve to "second". Instead, the following exception is thrown:

```
com.typesafe.config.ConfigException$BugOrBroken: replace in parent not possible ConfigDelayedMerge("first",${variable}) with Quoted("first") in ResolveSource(root=SimpleConfigObject({"variable":"second"}), pathFromRoot=null)
```

The following code blocks work as expected, so this is definitely an edge case that involves both an overloaded value AND resolving against another config:

```java
// This works
Config conf = ConfigFactory.parseString("{ \"one\": \"first\", \"one\": ${variable}, \"variable\": \"second\"}")
        .resolve();
```

```java
// This works
Config conf = ConfigFactory.parseString("{ \"one\": ${variable}}")
        .resolveWith(ConfigFactory.parseString("{ \"variable\": \"second\"}"));
```

Bug is observed running with artifact `com.typesafe:config:1.4.9`.

arimu1:
## Summary
- Fix `ConfigException.BugOrBroken` when calling `resolveWith()` on a config that has duplicate keys where the later value is a substitution (creating a `ConfigDelayedMerge`).
- Track the parent chain for configs being resolved even when the substitution lookup root is a different object (as `resolveWith()` does).
- Preserve the external substitution root during delayed-merge parent replacements so `${variable}` still resolves from the `resolveWith` source.

Reproduces and fixes https://github.com/lightbend/config/issues/855:

```java
Config conf = ConfigFactory.parseString("{ \"one\": \"first\", \"one\": ${variable}}")
    .resolveWith(ConfigFactory.parseString("{ \"variable\": \"second\"}"));
// expected: conf.getString("one") == "second"
```

## Test plan
- [x] Added `resolveWithDelayedMergeOverloadedValue` regression test in `ConfigTest`
- [x] `sbt config/test` — 553 tests passed (JDK 17)

Fixes #855

Made with [Cursor](https://cursor.com)

### DIFF
--- a/config/src/main/java/com/typesafe/config/impl/ResolveSource.java
+++ b/config/src/main/java/com/typesafe/config/impl/ResolveSource.java
@@ -167,7 +167,20 @@ final class ResolveSource {
             ConfigImpl.trace("pushing parent " + parent + " ==root " + (parent == root) + " onto " + this);
 
         if (pathFromRoot == null) {
-            return new ResolveSource(root, new Node<Container>(parent));
+            if (parent == root) {
+                return new ResolveSource(root, new Node<Container>(parent));
+            } else {
+                if (ConfigImpl.traceSubstitutionsEnabled()) {
+                    // this hasDescendant check is super-expensive so it's a
+                    // trace message rather than an assertion
+                    if (root.hasDescendant((AbstractConfigValue) parent))
+                        ConfigImpl.trace("***** BUG ***** tried to push parent " + parent
+                                + " without having a path to it in " + this);
+                }
+                // ignore parents if we aren't proceeding from the
+                // root
+                return this;
+            }
         } else {
             Container parentParent = pathFromRoot.head();
             if (ConfigImpl.traceSubstitutionsEnabled()) {
@@ -267,14 +280,19 @@ final class ResolveSource {
             Container parent = pathFromRoot.head();
             AbstractConfigValue newParent = parent.replaceChild(old, replacement);
             return replaceCurrentParent(parent, (newParent instanceof Container) ? (Container) newParent : null);
-        } else {
-            if (old == root && replacement instanceof Container) {
+        } else if (old == root) {
+            if (replacement instanceof Container) {
                 return new ResolveSource(rootMustBeObj((Container) replacement));
             } else {
                 throw new ConfigException.BugOrBroken("replace in parent not possible " + old + " with " + replacement
                         + " in " + this);
-                // return this;
             }
+        } else {
+            // pathFromRoot is null so we aren't proceeding from the root (see
+            // pushParent); resolveWith() resolves a value that isn't a descendant
+            // of the substitution root, so root can't contain "old" and there's
+            // nothing to replace.
+            return this;
         }
     }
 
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @arimu1,

Before we proceed with reviewing this pull request, please sign the Akka Contributors License Agreement:

[https://contribute.akka.io/contribute/cla](https://contribute.akka.io/contribute/cla)

config/src/test/scala/com/typesafe/config/impl/ConfigTest.scala
```diff
@@ -1387,6 +1387,15 @@ class ConfigTest extends TestUtils {
         assertEquals(43, resolved.getInt("foo"))
     }
 
+    // https://github.com/lightbend/config/issues/855
+    @Test
+    def resolveWithDelayedMergeOverloadedValue(): Unit = {
```
johanandren: I think this would be a more descriptive/correct test name

```suggestion
    def resolveWithDuplicateKeyAndSubstitution(): Unit = {
```

config/src/main/java/com/typesafe/config/impl/ResolveSource.java
johanandren: Put the fix here instead, splitting the combined `old == root && replacement instanceof Container condition` is what makes the two cases distinguishable: `old == root` keeps main's behaviour exactly, including the throw when the replacement isn't a `Container`; only "not a descendant of root at all" becomes a no-op.

```java
  } else if (old == root) {
    if (replacement instanceof Container) {
      return new ResolveSource(rootMustBeObj((Container) replacement));
    } else {
      throw new ConfigException.BugOrBroken("replace in parent not possible " + old + " with " + replacement + " in " + this);
    }
  } else {
    // pathFromRoot is null so we aren't proceeding from the root (see
    // pushParent); resolveWith() resolves a value that isn't a descendant
    // of the substitution root, so root can't contain "old" and there's
    // nothing to replace.
    return this;
  }
```

The comment is doing the : `replaceWithinCurrentParent` exists so a delayed merge can hide itself from its own substitutions, and when the lookup root doesn't contain the merge stack there is nothing to hide.

johanandren (commented):
Thanks for tracking this down, the diagnosis is right, and the behaviour the PR produces is right too.

The main feedback is to move the fix down to the branch that actually throws, that will be twice as fast and it leaves pushParent's documented invariant and its ***** BUG ***** trace in place.

Also, we'd need a signed CLA to be able to merge this.

arimu1:
Addressed the review feedback in 30c2252:

- Restored `pushParent` invariant and `***** BUG *****` trace
- Moved the no-op fix to `replaceWithinCurrentParent` (only non-descendant case returns `this`; `old == root` keeps main's throw behavior)
- Renamed test to `resolveWithDuplicateKeyAndSubstitution`

`sbt config/test` — 553 passed.

CLA still needs to be signed by the human contributor before merge.

